PR #1701, opened 2021-05-30
[WIP] macOS support

michaelsbradleyjr:
This is my attempt to catch up @grendello's fork's `macos` branch with this repo's current `master`, cf. PR #1280.

Note: I'm not a C programmer (will learn as I go) and only in the last year+ have been getting familiar with "systems" programming, Makefiles, cmake, etc., after spending many years working with langs/envs such as Node.js, Clojure, etc. In other words, my changes in this draft PR don't represent a deep analysis or contribution, I mainly just fiddled with @grendello's modifications until I got them working again.

With these changes in place and a fresh clone (and on this PR branch), the following works on macOS (I'm using 10.15), as long as the necessaries have been installed with `brew`:
```
$ mkdir build && cd build
$ cmake -DCMAKE_BUILD_TYPE=Release -DUSE_MULTIMEDIA=ffmpeg ..
$ make
```

After building, I've tried running `build/notcurses-demo` in iTerm2 (`TERM=xterm-256color`), Alacritty (`TERM=alacritty`), and Kitty (`TERM=xterm-kitty`) on macOS (`COLORTERM=truecolor`):
```
$ build/notcurses-demo -p ./data
```

In all cases the demo doesn't run all the way through, always ends like this:
```
...
14│ animate│ 28.43ms│     11│ 123.13Ki│  386.9│ 7│ 1│71│ 482.62║FAILED
15│    reel│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
16│whiteout│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
17│uniblock│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
18│    view│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
19│   luigi│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
20│ sliders│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
21│ fallin'│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
22│  jungle│   0.00s│      0│     0.00│    0.0│ 0│ 0│ 0│   0.00║SKIPPED
[...]

PR #1731, opened 2021-06-06
github actions workflow: macos, ubuntu, windows+msys2

michaelsbradleyjr:
This PR provides a [GitHub Actions workflow](https://github.com/michaelsbradleyjr/notcurses/blob/github-actions-workflow/.github/workflows/tests.yml) along the lines of what was discussed in #1701. If you're interested in adding this workflow but want to see some changes beforehand, let me know and I'll be happy to make them.

As currently configured, it would run the workflow on every commit to any open PR and every commit to the `master` branch. If desirable, it could be changed to run the workflow on every commit to any branch, whether there is a PR open for it or not.

Because GitHub Actions only "activates" once a workflow has been added to a repo's `master` branch, I made a [temporary hard fork](https://github.com/michaelsbradleyjr/thf-nc#readme) and intend to delete it once the workflow lands in this repo's `master`, or you decide you don't want the workflow after all.

### Ubuntu

Everything would run as expected; see the [latest **ubuntu** run on the hard fork's `master` branch](https://github.com/michaelsbradleyjr/thf-nc/runs/2758916749?check_suite_focus=true).

### macOS

The run would fail during the `cmake` step; see the [latest **macos** run on the hard fork's `master` branch](https://github.com/michaelsbradleyjr/thf-nc/runs/2758916736?check_suite_focus=true).

However, if the changes in #1701 were included (plus a few small changes since last week), then the `cmake` and `make` steps would succeed, but there would be a failure in the `ctest` step; see the [latest **macos** run on the hard fork's `macos` branch](https://github.com/michaelsbradleyjr/thf-nc/runs/2758918196?check_suite_focus=true).

Note: I've already rebased #1701 to include the commit in this PR's branch and the small changes in the `macos` branch of the hard fork.

### Windows

(relevant to #463)

The run would fail during the `make` step; see the [latest **windows** run on the hard fork's `master` branch](https://github.com/michaelsbradleyjr/thf-nc/runs/2758916758?check_suite_focus=true).

Here harder problems are faced, though let me start from the bright side 🌞  ...

If you can successfully compile for MinGW-w64 in an MSYS2 context and you do a `-static` build, then the executable can be run apart from any special context, i.e. could be launched and run from `cmd.exe` or PowerShell in a suitable terminal emulator, without the user needing to know anything about nor having installed anything related to MSYS2. There's a bit of magicality about the result, but there are challenges in reaching it...

Note the `make` step fails with:
```
src/lib/internal.h:22:10: fatal error: termios.h: No such file or directory
   22 | #include <termios.h>
      |          ^~~~~~~~~~~
compilation terminated.
```

It has to do with the limits of MSYS2 with respect to POSIX compatibility. Perhaps there's hope, though? Maybe a project like [veeso/termiWin](https://github.com/veeso/termiWin) could help fill in the missing pieces, or maybe there's something even better and I'm simply not aware of it yet (I haven't tried hacking in `termiWin` just yet)?

## General notes

If/when the steps on macOS and Windows are able to proceed past `ctest` and `make`, respectively, some changes would likely be needed in later steps, since those steps were written to work on Linux/Ubuntu specifically, but I don't think those changes would be difficult to implement and so (in my mind) don't present a blocker for this PR.


dankamongmen:
this is the best thing to happen all year. go go go open source!!!!! @michaelsbradleyjr  i am in your debt

WSLUser:
For Windows, there's a few things this project needs to do to properly compile, let alone run but inclusion of the Windows SDK with #WIN32 LEAN AND MEAN (something like that) to be defined in the header of windows.h. I can take a look at this PR and see what changes may be able to be made to at least ensure the correct steps are performed but I'm no expert at build steps and usually just copy from the github examples and tweaking them from there.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,171 @@
...
+          brew install \
+            coreutils \
+            doctest \
+            ffmpeg \
+            libunistring \
+            ncurses \
+            pandoc \
+            readline
+
+      - name: Install tools and libraries via MSYS2
```
WSLUser, line 76: It's great to have a method for MSYS2 but this is not the same as native Windows compilation. Ideally this project can be used in Cygwin using these build steps here but also want something more native that terminals like ConEmu (which is a Windows only terminal) can easily use.

WSLUser: It would be great if you created another build process under this one for native Windows compilation. It should use "windows-latest" for the platform. I think GH does some stuff behind the scenes with that defined to understand the Windows SDK is needed.

michaelsbradleyjr: > It's great to have a method for MSYS2 but this is not the same as native Windows compilation

I don't think I understand what you mean, though there are a lot of things about compilers, etc. I don't understand because I got a pretty late start with those things after so many years with JS and whatnot.

I am currently using an MSYS2 / MINGW-w64 approach to build executables on Windows 10; having passed `-static` to the compiler I am able to copy the standalone executables to Windows 7 x64 or Windows 10 x64 machines that have nothing at all installed related to MSYS2, MINGW, nor Cygwin. And my executables run just fine in `cmd.exe`. Depending on what the executable does, it can be better (or practically necessary, e.g. if it's printing colored output) to run it in ConEmu or Windows Terminal (running cmd.exe) instead of firing up `cmd.exe` directly.

So, it's not native compilation in the sense that it doesn't use the MS compiler toolchain, but the end result, albeit involving a `-static` compile, is producing a native executable that doesn't depend on anything that doesn't ship with Windows.

I've dabbled with other ways of cross-platform building (for Linux, macOS, and Windows) when it comes to Windows, and I get the most "bang" for my time on Windows by opting for an MSYS2 build context and `-static` compilation.

That being said, I'll repeat that maybe I didn't understand your point. Also, I realize not everyone wants to use MSYS2 for the build context, so perhaps more measures are needed than just getting it to build that way, i.e. this repo's codebase and `CMakeLists.txt` could support building with the MS tooolchain, and other ways too. When it comes to that, I'll be happy to learn and help as I can, but I live in Emacs and bash shells; I've never used Visual Studio and am not really sure where to start when it comes to stuff that's not Unix-like in the way that MSYS2 is. Hopefully other community members can contribute to making this a thing on Windows in multiple best ways possible.

> It should use "windows-latest" for the platform.

Fwiw, the workflow I put together is already using "windows-latest": https://github.com/michaelsbradleyjr/notcurses/blob/github-actions-workflow/.github/workflows/tests.yml#L35

michaelsbradleyjr:
> For Windows, there's a few things this project needs to do to properly compile, let alone run such as the inclusion of the Windows SDK with #WIN32 LEAN AND MEAN (something like that) to be defined in the header of windows.h. I can take a look at this PR and see what changes may be able to be made to at least ensure the correct steps are performed but I'm no expert at build steps and usually just copy from the github examples and tweaking them from there.

I agree. There are likely multiple compiler and linker flags and headers and such that need to be added in the case of Windows before `make install` can succeed.

WSLUser:
Since we're talking about a TUI library here getting used on Windows (where it's currently not supported), I'll point to example of Windows Terminal (where I would love to see notcurses applications working regardless of WSL bash or Powershell) for their build stuff. Some of it is definetly specific to them but hopefully shows you what I mean about having a native Windows build vs a build using MingW/Msys2: https://github.com/microsoft/terminal/tree/main/build

michaelsbradleyjr:
> Since we're talking about a TUI library here getting used on Windows (where it's currently not supported) ... (where I would love to see notcurses applications working regardless of WSL bash or Powershell)

I'll take a look, and thank you for the link.

Again, though, I have in mind the library being usable to build TUI apps that run on Windows without *users* of those apps needing any libraries/environment that doesn't ship with Windows (even going back to Windows 7), in effect a native build.

How would users know, or why would they care, what compiler, linker, headers, C archives, etc. were used to create the executable? As I expressed previously, I do get it that some *developers* will have a preferred way of building on Windows that doesn't involve e.g. MSYS2, makes perfect sense.

So I still feel I'm missing an important point/s here, thanks for your patience.

Also, I'm currently working on something that involves ncurses (would prefer notcurses, of course). When I'm able to do show something, I will. Or maybe it's not going to work out like I thought and your point will suddenly become very clear to me. 😄

WSLUser:
So I would forget about Win7 support. I know there's many applications that haven't dropped it yet but this project isn't even available for Windows yet. Any new applications or libraries brought to Windows should consider only modern Windows. In terms of maintenance and optimizations, trying to support WIn7 just isn't worth it. We should be encouraging users to go to 10, not stick with EOL versions. We can support 8.1 since that is fully supported for now but I really think we should target newer versions of Win10 that would make it far easier for a library that currently supports Linux than do many of the trickeries that applications such as the Cygwin terminal Mintty (which doesn't even rely on the underlying conhost) do in order to support downlevel OSs. Lastly, we oughtn't really even support Windows until conhost does get Sixel implemented at least. Otherwise, only terminals like Mintty will actually be able to utilize this library but far more terminals rely on conhost and ConPTY than don't. But don't let that stop you from ensuring a successful build can't be completed. We can have all the preparation done so that when the support is there, we can release is relatively no time.

dankamongmen:
alright, sorry for the delay in reviewing this. on it now!

dankamongmen:
amazing. and pleasantly self-contained. in it goes! i look forward to a weekend of Win32 hackery!

dankamongmen:
making you a collaborator, @michaelsbradleyjr , or sending an invite at least. you seem tremendously useful.

michaelsbradleyjr:
I'm not sure why yet, but the GitHub Actions workflow wasn't automatically activated for this repo when you merged my PR.

However, it is [running for my fork](https://github.com/michaelsbradleyjr/notcurses/actions)... strange... or maybe not and there's a simple explanation, just I'm not aware of it yet. I'm currently researching the problem; see also https://github.com/michaelsbradleyjr/notcurses/pull/1.


Note that I've deleted my temporary hard fork, since it's no longer necessary.

michaelsbradleyjr:
@dankamongmen, since you're admin level for this repo can you...

(1) See if you can navigate to https://github.com/dankamongmen/rln/actions or https://github.com/dankamongmen/rln/actions/new

(2) If the previous step works, is there an option to enable the Tests workflow that was made available by this merged PR?

michaelsbradleyjr:
Might be relevant: https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#managing-github-actions-permissions-for-your-repository

It may be that you need to activate GitHub Actions in the settings for this repo and/or the settings of your personal GitHub account.

igo95862:
@michaelsbradleyjr github actions got enabled.

I think MacOS and Windows builds should be optional until they are officially supported otherwise it will cause normalization of failure by its constant failures.

michaelsbradleyjr:
> github actions got enabled.

Nice.

> I think MacOS and Windows builds should be optional until they are officially supported otherwise it will cause normalization of failure by its constant failures.

"Normalization of failure" would not be good. On the other hand, I thought the "red lights" in GitHub Actions could be a bit of a motivator to get `make`, `ctest`, etc. working on all three platforms sooner rather than later. 

> Do you mind if I edit the actions a bit.

Not at all, that's what open source and pull requests are all about, and PRs are a good context for discussing proposed changes as well, since now that GHA is enabled for the repo, PRs that modify the GHA workflow will run with the modified workflow.

> (plus add my python binds build)

Are those different from the Python bindings that are already part of the GHA workflow?

michaelsbradleyjr:
> github actions got enabled.

@igo95862, it seems GHA isn't fully activated; see the following screenshot taken from the bottom of: https://github.com/dankamongmen/notcurses/actions/runs/934132664

image

That might be deliberate, or maybe the needed settings changes were overlooked or not understood.

The workflow I put together makes use of two popular and well-trusted GitHub Actions for (1) setting up MSYS2 in runs on Windows, and (2) checking out the source code on all platforms. Allowing a workflow to use 3rd party actions would need to be toggled in this repo's settings.

Also note that after GHA was generally flipped (back) on, there's now a very old workflow (named CMake) from about 9 months ago that gets listed in the Actions top-level view: https://github.com/dankamongmen/notcurses/actions/workflows/cmake.yml. Deleting that branch from this repo should stop it (and its failed runs) from being displayed.

igo95862:
> @igo95862, it seems GHA isn't fully activated; see the following screenshot taken from the bottom of: https://github.com/dankamongmen/notcurses/actions/runs/934132664

Yes, @dankamongmen has only enabled local actions. 

Screenshot from 2021-06-14 08-44-10

@dankamongmen do you mind enabling all or verified actions?

PR #1920, opened 2021-07-11
refactor FAQ section for easier navigation

joseluis:
This is an attempt of making the FAQ section on the README easier to consult, less lengthy, and prettier for the users.

This changes the markdown format for HTML tags, but since it doesn't have to be edited frequently, I don't think it's gonna become a problem.


joseluis:
before:
image

after:
image

dankamongmen (approved):
does `tt` not work in plane of `code`? i guess it's less semantic, ok.

PR #2513, opened 2021-12-31
Restore iTerm2 image code (wip)

klamonte:
For #2077

iTerm2 image code restored to a compilable state.  I made some guesses on tracking the internal API changes.  I also enabled iTerm2 for WezTerm and mintty.

The remaining problems:

1. Failed to link with libdeflate.  `cmake -DUSE_DEFLATE=OFF ..` made it use libz and it compiled OK.

2. It segfaults in the demo.

Hoping someone can point me in the right direction to fixing the segfault.  I would really like to see the demo with PNGs and alpha transparency. :-)

### DIFF
--- a/src/lib/iterm.c
+++ b/src/lib/iterm.c
@@ -1,0 +1,101 @@
+// the iterm2 graphics protocol is based entirely around containerized formats
+// https://iterm2.com/documentation-images.html
+
+#include <stdio.h>
+#include "internal.h"
+#include "termdesc.h"
+#include "sprite.h"
+#include "fbuf.h"
+#include "png.h"
+
+// yank a cell out of the PNG by setting all of its alphas to 0. the alphas
+// will be preserved in the auxvec.
+int iterm_wipe(sprixel* s, int ycell, int xcell){
+  (void)s;     // FIXME
+  (void)ycell;
+  (void)xcell;
+  return -1;
+}
+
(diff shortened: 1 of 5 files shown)


klamonte:
I think I've changed my mind.

iTerm2 PNG gives one the ability to put pixels on the bottom row, IF the terminal has implemented "doNotMoveCursor", and so far only wezterm does that.  So the only things you gain (visually) with iTerm2 output is true alpha-blending and the 24-bit color depth.  But a sixel encoder using smaller image fragments can approach 16- to 19.97-bit color depth anyway.  Add an alpha cutoff to make mostly-transparent pixels fully transparent, do image-over-image blending during layer compositing, and it might be hard in practice to even see the difference between sixel and PNG.

Where iTerm2 is a clear win is in bandwidth: PNG and JPG are a _lot_ smaller than sixel for the same number of pixels.  But then only three of the terminals that can read iTerm2 (mintty, wezterm, jexer) can handle higher throughput images anyway -- unless iTerm2 itself has had improvements on that front in the last 18ish months.

@dankamongmen Do you have a preference?  Should I close this out at this time, or try to get it working?

klamonte:
Furthermore, sixel just [might be able to emit to the bottom row afterall.](https://gitlab.com/klamonte/jexer/-/issues/91)   So even the wezterm-only doNotMoveCursor option with iTerm2 might not give it a real edge.

PR #2598, opened 2022-02-04
Use the right code for unsigned int.

alexhsamuel:
This fixes a core dump in `003-color-example.py`.


dankamongmen:
thanks very much!

PR #2681, opened 2022-09-25
Fixes DECSDM interpretation for Contour.

christianparpart:
We have that fixed since https://github.com/contour-terminal/contour/pull/431 (October 2021) - I don't see why we should `if`-guard here, either. So simply flipping the bit.

### DIFF
--- a/src/lib/termdesc.c
+++ b/src/lib/termdesc.c
@@ -846,7 +846,7 @@ apply_contour_heuristics(tinfo* ti, size_t* tablelen, size_t* tableused,
   ti->caps.sextants = true;
   ti->caps.rgb = true;
   *forcesdm = true;
-  *invertsixel = true;
+  *invertsixel = false;
   return "Contour";
 }
 


dankamongmen:
looks good to me, thanks!

dankamongmen (approved):

dankamongmen:
FWIW, i normally guard this kind of thing with a version check. was the version that inverted DECSDM released in any distributions? it was in arch, right?

christianparpart:
> FWIW, i normally guard this kind of thing with a version check. was the version that inverted DECSDM released in any distributions? it was in arch, right?

Despite its functionality, I don't think Contour is as famous as being included downstream in any distribution just yet. We have an Arch AUR (but not under my control) and we deploy all the releases via Github release page plus recently Flathub, which makes updating super convenient. I was thinking of adding that guard (hence my statement on top) but I decided against it because the fix is in since so long already (11 months) and most users I am in contact with use master builds or are up to date with Github release page builds. :)

dankamongmen:
if you're not worried about it, i'm not worried about it! hope you're doing well, friend

christianparpart:
Yeah, thanks. And the more we go into the future the more likely I need to
object on guards. Absolutely.

Have a look at the discussion I posted this afternoon. I am curious about
your feedback.

PR #2698, opened 2023-01-23
Speed up startup that got slow since #2696

### LINKED ISSUE #2697: Starting up a notcurses program takes a (short, but still) while
I'm writing a program using notcurses that I'd like to have start up instantly. Unfortunately, at the moment and with my hardware (a 5-year-old Core-I7 ULP), that is not what's happening: it takes about a second for notcurses_init() to finish. Profiling leads to the add_phi_and_eta_recurse call taking the majority of the time in this call; that seems to be part of the input escape code automaton.

I have no issue looking into this myself, but @dankamongmen , do you happen to have an ideas how this function can be optimized?

Spritetm:
#2696 added a bunch of numeric fields to the input automaton, greatly slowing down startup times of any notcurses program. This is mostly because numeric fields aren't handled optimally in the automaton: on creation, a lot of links are created from one trie node via the digits 0-9 to another one. When appending more patterns, the automaton creation logic takes each of these digit links as an individual path that should be followed recursively, which greatly increases processing time. This patch makes the logic recognize this and skips following a path if it is the same as the path followed before.

To illustrate, the times on my laptop to run notcurses-info are as following:
Master, before #2696: 0m0.031s                                                                                 
Master, after #2696: 0m1.001s                                                                                      
After this patch: 0m0.034s  

Closes #2697

### DIFF
--- a/src/input/input.cpp
+++ b/src/input/input.cpp
@@ -361,6 +361,16 @@ int input_demo(ncpp::NotCurses* nc) {
         n->printf("Unicode: [0x%08x] '%s'", r, ni.utf8);
       }
     }
+    if(ni.eff_text[0] != ni.id || ni.eff_text[1] != 0){
+      n->printf(" effective text '");
+      for (int c=0; ni.eff_text[c]!=0; c++){
+        unsigned char egc[5]={0};
+        if(notcurses_ucs32_to_utf8(&ni.eff_text[c], 1, egc, 4)>=0){
+          n->printf("%s", egc);
+        }
+      }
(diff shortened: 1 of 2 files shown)


Spritetm:
Sorry, hold this for a sec, I think I found a bug...

Spritetm:
Okay, fixed now. I broke my own code to get the effective text somehow :X Along with the fix is an addition to ncurses-input to show the effective text if it differs from the key id/utf8; this should make bugs that break #2696 easier to spot.

PR #2891, opened 2025-08-05
Fix manpages installation via Python wheels

hroncok:
All data files are relative to the prefix.
This should also fix the mentioned --user without --prefix case.

Without this, the manpages end up installed in site-packages.


dankamongmen:
thanks much!

PR #2893, opened 2025-08-16
avoid resizing bitmap when source meets the destination

mjkpolo:
This addresses the `FIXME` here:
https://github.com/dankamongmen/notcurses/blob/8edce88451f89544aec11e54278544532a7a98dd/src/lib/internal.h#L1736
I noticed some mild distortion when drawing a large bitmap with `NCBLIT_2x1`. This fixes my issue. It might not be very elegant so I'm open to any feedback!


mjkpolo:
If you need a more minimal example let me know, but here are some screenshots from my [repo](https://github.com/mjkpolo/dogan) with and without the change where green circles are how it's supposed to look, and red circles are what it looks like without the patch

with_patch
without_patch

The errors are from drawing [this png](https://github.com/mjkpolo/dogan/blob/main/gimp/water_border.png) so you can see the same distortion comparing to it

mjkpolo:
Let me know if there's anything else I can do to help get this merged, thanks!

Issue #2894, opened 2025-08-24
(Windows Msys2 UCRT64) build failure: "error: implicit declaration of function 'setenv';"


AmmoniumX:
Compilation was done following https://github.com/dankamongmen/notcurses/blob/master/INSTALL.md#microsoft-windows 

Reproduction Steps:
- Open Msys2 UCRT64, install build dependencies by running command on INSTALL.md:
`pacman -S mingw-w64-ucrt-x86_64-cmake mingw-w64-ucrt-x86_64-libdeflate mingw-w64-ucrt-x86_64-libunistring mingw-w64-ucrt-x86_64-ncurses mingw-w64-ucrt-x86_64-ninja mingw-w64-ucrt-x86_64-openimageio mingw-w64-ucrt-x86_64-toolchain`
- Clone notcurses, make build directory, and generate build configuration: `cmake .. -DUSE_DOCTEST=off -DUSE_PANDOC=off`
- Attempt building: `cmake --build .`

Resulting error:
```
diego@DGC-TUF-A15 UCRT64 ~/notcurses/build
$ cmake .. -DUSE_DOCTEST=off -DUSE_PANDOC=off
-- Building for: Ninja
-- The C compiler identification is GNU 15.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64/ucrt64/bin/cc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is GNU 15.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/msys64/ucrt64/bin/c++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Requested multimedia engine: ffmpeg
-- Requested build mode: RelWithDebInfo
-- Found PkgConfig: C:/msys64/ucrt64/bin/pkg-config.exe (found version "2.5.1")
-- Checking for one of the modules 'tinfo>=6.1;ncursesw>=6.1'
-- Checking for module 'libavcodec>=57.0'
--   Found libavcodec, version 61.19.101
-- Checking for module 'libavdevice>=57.0'
--   Found libavdevice, version 61.3.100
-- Checking for module 'libavformat>=57.0'
--   Found libavformat, version 61.7.100
-- Checking for module 'libavutil>=56.0'
--   Found libavutil, version 59.39.100
[...]

PR #2895, opened 2025-08-24
fix: setenv not defined on Windows UCRT

AmmoniumX:
Closes https://github.com/dankamongmen/notcurses/issues/2894 

This is just the simplest solution I could think of for the issue, let me know if you'd prefer a different approach

### DIFF
--- a/src/lib/termdesc.c
+++ b/src/lib/termdesc.c
@@ -8,8 +8,7 @@
 #include "windows.h"
 #include "linux.h"
 #ifdef _WIN32
-// On Windows, map setenv to the non-standard _putenv
-#define setenv(name, value, overwrite) _putenv_s(name, value)
+#define setenv(name, value, override) SetEnvironmentVariable(name, value)
 #endif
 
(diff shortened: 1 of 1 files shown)


dankamongmen:
hey just seeing this now, sorry!

how were things working before? has it been broken for a long time? ugh

let me look at this real quick, and i'll probably merge it. thanks for the patch!

dankamongmen (changes requested):
Please remove the stuff regarding `altstack`. It ought be `NULL` if not allocated, and thus safe to pass to `free()`. If that's not the case, it's a bug, and we ought fix it. I think we need `SetEnvironmentVariable()` instead of `putenv()`. See my comments for explanation. Thanks for the patch! Make those changes and I'll merge it.

dankamongmen:
ok, i think https://github.com/dankamongmen/notcurses/pull/2903 has a good fix for this. thank you for bringing it to my attention, and sorry for the delay! i've credited you in the PR with the report.

PR #2897, opened 2025-10-04
Don't `#include` within an `extern "C"` block

ryandesign:
Using `#include` within an `extern "C"` block can break the header being included. For example, build fails when using libunistring >= 1.4.

[libunistring 1.4 has a separate `extern "C"` bug](https://savannah.gnu.org/bugs/index.php?67576) to be fixed in 1.4.1. But once that's fixed, notcurses still fails with:

```
In file included from /opt/local/var/macports/build/notcurses-0a839c9e/work/notcurses-3.0.9/src/tests/Exceptions.cpp:2:
In file included from /opt/local/var/macports/build/notcurses-0a839c9e/work/notcurses-3.0.9/src/tests/main.h:13:
In file included from /opt/local/var/macports/build/notcurses-0a839c9e/work/notcurses-3.0.9/src/lib/internal.h:47:
In file included from /opt/local/var/macports/build/notcurses-0a839c9e/work/notcurses-3.0.9/src/lib/egcpool.h:13:
/opt/local/include/unigbrk.h:121:1: error: templates must have C++ linkage
template <typename T>
^~~~~~~~~~~~~~~~~~~~~
/opt/local/var/macports/build/notcurses-0a839c9e/work/notcurses-3.0.9/src/lib/internal.h:5:1: note: extern "C" language linkage specification begins here
extern "C" {
^
```

Fixing just src/lib/internal.h so that the `#include` statements aren't inside the `extern "C"` block fixes this build failure for me, but in this PR for good measure I've changed all of the files that did this.

### DIFF
--- a/include/notcurses/ncport.h
+++ b/include/notcurses/ncport.h
@@ -1,10 +1,6 @@
 #ifndef NOTCURSES_NCPORT
 #define NOTCURSES_NCPORT
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 // Platform-dependent preprocessor material (includes and definitions) needed
 // to compile against Notcurses. A critical definition is htole(), which forces
 // 32-bit values to little-endian (as used in the nccell gcluster field). This
@@ -34,10 +30,6 @@ extern "C" {
 #else                                             // BSDs
 #include <sys/endian.h>
 #define htole(x) (bswap32(htonl(x)))
-#endif
-
-#ifdef __cplusplus
-} // extern "C"
 #endif
 
 #endif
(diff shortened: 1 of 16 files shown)


dankamongmen:
let me take a closer look at this, but i'm pretty sure we need these guards for c++ linkage. thanks for the report and patch. i'll try to look at it tonight or tomorrow.

dankamongmen:
Yeah, I think the solution in @2902 is better. We definitely need these `extern "C"` linkage specifiers on our own functions, but shouldn't be applying them to external definitions. Closing this up, but your problem ought be solved.

Issue #2892, opened 2025-08-06
ncvisual_options used region not working


Ben0981:
Hello, I have been trying to use the ncvisual_options to crop a resized image but no matter what I do the `lenx` `leny` `begx` `begy` parameters seem to be ignored.

Here is my code:
```c
#include "notcurses/notcurses.h"

int main(int argc, char** argv) {

	notcurses_options opts = {};
	struct notcurses* nc = notcurses_init(&opts, NULL);

	struct ncplane* stdn = notcurses_stdplane(nc);

	struct ncvisual* img = ncvisual_from_file("image.png"); 

	ncvisual_resize(img, 400, 400);

	struct ncvisual_options imgOpts = {
		.y = 0, .x = 0,
		.begx = 0, .begy = 200,
		.lenx = 200, .leny = 200,
		.blitter = NCBLIT_PIXEL,
		.n = stdn,
		.flags = NCVISUAL_OPTION_CHILDPLANE,
	};

	struct ncplane* pixelPlane = ncvisual_blit(nc, img, &imgOpts);
	
	notcurses_render(nc);

	ncinput ni = {};
	notcurses_get_blocking(nc, &ni);

	notcurses_stop(nc);
	
	return 0;
}
```

What I would expect to happen here is a 200x200px subsection, showing the lower left quadrant, of the image being rendered, but what actually happens is the full 400x400px image being rendered (the same happens without the resize but larger). Setting these parameters to any other value also does not change what is rendered.

Looking at the man-pages my usage should be correct, with `begy` and `begx` setting the top left corner and `leny` and `lenx` setting the size of the rendered section.

I am not sure if I am using this wrong or if something isn't working but any help with this is much appreciated.

PR #2925, opened 2026-05-18
ncvisual_blit_internal: honor begy/begx/leny/lenx on every blit path

m-doughty:
## Resolves

- Issue #2892 
- `// FIXME begy/begx are really only of interest to scaling…` (src/lib/internal.h:blitterargs)
- `// FIXME need to honor leny/lenx and begy/begx` (src/media/oiio.cpp)

## Overview

I found this issue while doing some work in [Selkie::Widget::ViewportedCardList](https://github.com/m-doughty/Selkie/blob/master/lib/Selkie/Widget/ViewportedCardList.rakumod) then realised it was already marked by `FIXME`s and an issue upstream, so I decided it'd be better to fix it at source instead of in a local shim.

`ncvisual_blit_internal` takes `begx`, `begy`, `lenx` and `leny` (the source subregion in pixel space) but most downstream backends disregard them.

- **Sprixel blitters** take only `data`, `lenx` and `leny` and consume from `data[0]` so any crop with `begx` or `begy` > 0 renders the top left instead of the required region.
- **Cell blitters** reference `begx` and `begy` but index the resized buffer within the input-space offset -- latent out of bounds read for `begy` > 0 whenever the resize ratio shrinks rows.
- **OIIO blit** had an explicit `// FIXME need to honor leny/lenx and begy/begx`

## Fix Summary

- Crops the requested sub-region once at the entry of `ncvisual_blit_internal` then zeroes `begx`, `begy`, `lenx`, `leny` in a local `blitterargs` copy before dispatching to backends (backends now uniformly get the pre-cropped source).
- Adds a static helper to `src/lib/visual.c` called `ncvisual_blit_internal` which:
  - Allocates a fresh `ncvisual` and row-loop `memcpy`s the requested region honoring source `rowstride` padding
  - Re-pads the destination via `pad_for_image`
  - Calls `ncvisual_details_seed` so the cropped visual feeds through the same FFmpeg/OIIO pipeline
  - Preserves the original `barg` for callers, only the local copy sent to backends is zeroed
- Skipped when no crop is needed via quick short-circuit for the common full-source case

## Testing

Added `Visual::BlitSubregionOffset` (src/tests/visual.cpp):

- Builds an 8x8 RGBA source with four 4x4 quadrants in distinct colors
- Blits with `begx = 4`, `begy = 4` and asserts that every cell is returned white (bottom right quadrant)
- This test was red-green tested, failed on `master` and `v3.0.17` with 48/73 assertions reporting red

Full `notcurses-tester` suite green on both bases (MacOS aarch64, ffmpeg backend).

### DIFF
--- a/src/lib/internal.h
+++ b/src/lib/internal.h
@@ -370,8 +370,6 @@ typedef struct notcurses {
 } notcurses;
 
 typedef struct blitterargs {
-  // FIXME begy/begx are really only of interest to scaling; they ought be
-  // consumed there, and blitters ought always work with the scaled output.
   int begy;            // upper left start within visual
   int begx;
   int leny;            // number of source pixels to use
--- a/src/lib/visual.c
+++ b/src/lib/visual.c
@@ -81,29 +81,71 @@ ncplane* ncvisual_subtitle_plane(ncplane* parent, const ncvisual* ncv){
   return visual_implementation->visual_subtitle(parent, ncv);
 }
 
+// Forward decl; definition needs pad_for_image in scope (below).
+static ncvisual* ncvisual_subregion_internal(const ncvisual* src,
+                                             unsigned begy, unsigned begx,
+                                             unsigned leny, unsigned lenx);
+
 int ncvisual_blit_internal(const ncvisual* ncv, int rows, int cols, ncplane* n,
                            const struct blitset* bset, const blitterargs* barg){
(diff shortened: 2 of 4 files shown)


m-doughty:
@dankamongmen Did you get a chance to look at this? Not sure if it's how you'd prefer to deal with this issue or not. Happy to rework it if you had a different direction in mind.

PR #2927, opened 2026-05-22
Add support for utf8proc as an alternative to libunistring

stefan-sedlak:
This adds optional `utf8proc` support as an alternative Unicode backend to GNU `libunistring`.

## Motivation

The main motivation is licensing.

`notcurses` currently depends on `libunistring` for a small set of Unicode operations. That works well technically, but `libunistring` is LGPL-licensed, which is a blocker for some downstream static-linking use cases. In my case, I need to statically link `notcurses` into a larger project, and the current `libunistring` dependency makes that difficult to ship under my project's licensing constraints.

`utf8proc` is a permissively licensed alternative (MIT/Expat-style license) with no nested runtime dependencies, and it provides the pieces needed here: grapheme breaking, UTF-8 encoding, and Unicode category lookup.

This change is intended to make that use case possible without changing the default behavior for existing users.

## What this changes

- adds a new CMake option: `USE_UTF8PROC`
- keeps `libunistring` as the default backend
- adds a small internal Unicode abstraction layer so backend-specific logic stays localized
- switches the existing Unicode-dependent call sites to that abstraction
- updates generated `pkg-config` metadata so the selected backend is reflected correctly
- updates the Unicode regression test wording to be backend-neutral
- documents the new optional backend in the README

## Why this shape

The goal here was to keep the patch small and low-risk for upstream review:

- default behavior is unchanged
- no automatic backend switching or preference logic was added
- the existing `libunistring` path remains intact
- backend-specific code is isolated instead of being spread across the tree

In other words, this is an additive change, not a backend replacement.

## Notes on behavior

This patch is not trying to redesign `notcurses` Unicode handling. It only introduces an alternative implementation for the existing dependency surface.

The current `libunistring` backend remains the default and should behave exactly as before. `utf8proc` is opt-in via `-DUSE_UTF8PROC=ON`.

## Validation

I validated the following:

- default build with `libunistring`
- build with `-DUSE_UTF8PROC=ON`
- build of `notcurses-tester`
- generated `pkg-config` output for both backend selections

## Why this is useful

This gives downstream users a permissive licensing path for static linking while preserving the current default for existing packagers and users of `libunistring`.

If this approach is acceptable, it should make `notcurses` easier to integrate into projects that need a more permissive dependency story without forcing any change on current consumers.


stefan-sedlak:
Hi @dankamongmen , is it something useful for the library or shall I close the pull request?

PR #2933, opened 2026-08-03
egcpool: clear injoin after a ZWJ-joined pictographic (GB11)

btpence:
utf8_egc_len() set `injoin` on the first ZWJ and never cleared it. Since the grapheme-break check is gated on `!injoin`, once a ZWJ was seen the function stopped breaking and folded every following EGC into the cluster (each at zero columns), so ncstrwidth() under-counted.

Per GB11 a ZWJ suppresses the break before exactly one pictographic. Reset injoin after that codepoint; multi-part sequences re-arm on each ZWJ.

Repro (returned 2, want 4):
  ncstrwidth("😮‍💨")   == 2
  ncstrwidth("🤥")     == 2
  ncstrwidth("😮‍💨🤥") == 2   // bug

Add MultiglyphWidth regression tests.

also fixed utf8_egc_len....

utf8_egc_len() ended the grapheme at a variation selector. VS15/VS16 are
Grapheme_Extend, so this orphaned a following combining mark into its own
zero-width EGC.


btpence:
Fixes incorrect space calculation for emoji.
Before:
image
After:
image

PR #2937, opened 2026-08-07
input: accept the DA1 reply sixel-enabled tmux sends

nettrino:
### Problem

Inside a sixel-enabled tmux (brew's `tmux` 3.5a, `--enable-sixel`), every
notcurses program hangs at startup on a blank screen — `notcurses-info`
included. It never returns from `notcurses_init` → `interrogate_terminfo`.

That tmux answers a Primary Device Attributes query with `CSI ? 1 ; 2 ; 4 c`:

```
$ printf '\033[c' ; # inside tmux 3.5a
^[[?1;2;4c
```

`build_cflow_automaton` registers the literal `"[?1;2c"`, and that literal
branch shadows the `"[?\N;\Dc"` wildcard that would otherwise absorb the
trailing `;4`. The walk dies partway through the reply, `da1_cb` never fires,
and interrogation waits forever for a DA1 it already received.

### Fix

Register the exact string tmux sends, next to the other literal DA1 forms.

### Notes

Reproduced against both v3.0.17 and master, on macOS 26 / arm64. Verified by
building master with the change and running under the same tmux.

Happy to add a unit test if you'd like one — as far as I can tell the cflow
table is private to `in.c` and `build_cflow_automaton` isn't reachable from
`src/tests`, so testing it would mean exposing the table (or the builder), and
I didn't want to make that call for you in a one-line bugfix.

We're carrying this as a downstream patch in the meantime; happy to adjust the
comment style or the placement if you'd prefer it handled through the wildcard
instead.


nettrino:
Prior art worth flagging: this is the same reply reported in #2255 (2021, rxvt-unicode-sixel), where the suggested fix was the identical `{ "[?1;2;4c", da1_cb, },` entry. That issue was closed in 2021, but the entry isn't in the table on master today — so either it never landed or it was lost in a later automaton rewrite. What's changed since is the blast radius: sixel-enabled tmux 3.5a sends the same DA1 reply, so this now hits a fairly common setup rather than one patched terminal fork.

j4james:
See also issue #2736, and PRs #2908 and #2926.

